PR #9, opened 2016-05-18
Update to 1.61

jschueller:


msarahan:
The Windows logs are enormous because files are copied one-by-one.  I didn't have the patience to wait for my browser to load the whole log to find the error.  I recommend changing this recipe, potentially bypassing their built-in logic, and copying or (better) moving directories rather than individual files.

jschueller:
@msarahan, you can download the raw log file instead, it's a problem because of an hardcoded path.

jakirkham:
There still seems to be some path issue. I looked at the log, but it is pretty vague about the error.

jschueller:
@jakirkham great success!

jakirkham:
What was causing the problem?

recipe/bld.bat
```diff
@@ -33,11 +33,11 @@ if errorlevel 1 exit 1
 if errorlevel 1 exit 1
 
 :: Install fix-up for a non version-specific boost include
-move %LIBRARY_INC%\boost-1_60\boost %LIBRARY_INC%
+move %LIBRARY_INC%\boost-1_61\boost %LIBRARY_INC%
 if errorlevel 1 exit 1
 
 :: Move dll's to LIBRARY_BIN
-move %LIBRARY_LIB%\*vc%VSTRING%0-mt-1_60.dll "%LIBRARY_BIN%"
+move %LIBRARY_LIB%\*vc%VSTRING%0-mt-1_61.dll "%LIBRARY_BIN%"
```
jakirkham: Now that this appears to be a persistent feature, we should probably absorb it somehow. We probably can do something similar to [this]( https://github.com/conda-forge/tk-feedstock/blob/ceeb3993e637d82cbac4c131d6a477282c6ef2b3/recipe/bld.bat#L8-L9 ).

jschueller:
@jakirkham ok, applied

ccordoba12:
Great job @jschueller!!

jakirkham:
Awesome @jschueller! I would like to merge this soon, but I realized earlier we are not pinning `boost`. If you guys could help me track down and pin `boost` to `1.60.x` on packages that are using it, I would really appreciate it. This is not to say we won't upgrade to `1.61.x`. Just that I'm worried about breaking people's existing stuff. I'm going to add `WIP` here until we have finished that.

jschueller:
@jakirkham
https://github.com/search?l=yaml&q=org%3Aconda-forge+%22-+boost%22&type=Code

jakirkham:
I'm aware. Just help from others doing this pinning will be helpful as this is only one of many things that needs to get done.

jschueller:
@jakirkham 
https://github.com/conda-forge/caffe-feedstock/pull/6
https://github.com/conda-forge/bob.extension-feedstock/pull/6
https://github.com/conda-forge/thrift-cpp-feedstock/pull/5
https://github.com/conda-forge/eccodes-feedstock/pull/5
https://github.com/conda-forge/libradolan-feedstock/pull/2
https://github.com/conda-forge/rank_filter-feedstock/pull/5
https://github.com/conda-forge/bob.math-feedstock/pull/6
https://github.com/conda-forge/bob.blitz-feedstock/pull/6
https://github.com/conda-forge/vigra-feedstock/pull/10

jakirkham:
Thanks @jschueller. Couple points.

* Build numbers must be bumped.
* Let's do `1.60.*` with that last `.`.

Otherwise, I'll merge them after this is changed.

jakirkham:
@ocefpaf @183amir, @kayhayen, if you are not aware, we are planning on releasing Boost 1.61.0. As this may contain ABI breaking changes, we have tried to track down feedstocks using Boost and add pinning PRs to fix them to `1.60.*` (current). All of this has been done in no small part by @jschueller. If you have no strong objections and CI is passing, please merge the pinnings. I'll try to check back in on Monday and merge any stragglers so that we don't break your stacks when we release this. For the broken ones, we should probably figure out what is wrong and see how we can address it.

jakirkham:
Also, @jschueller, please feel free to add yourself to the maintainers list here.

kayhayen:
Hello there,

Nuitka does not depend  on Boost at all. So green light from me here.

Yours,
Kay

jakirkham:
Everything is pinned that needs to be so let's get `1.61.*` out.

PR #12, opened 2016-06-24
Point-independent code to avoid static linking issues

wesm:
The point independent code means that the boost static libraries can be properly statically linked while creating other shared libraries. This makes so that users of those shared libraries do not need to install the boost dynamic libraries on their system.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -13,6 +13,9 @@ set -x -e
 INCLUDE_PATH="${PREFIX}/include"
 LIBRARY_PATH="${PREFIX}/lib"
 
+# Always build PIC code for enable static linking into other shared libraries
+CXXFLAGS="${CXXFLAGS} -fPIC"
+
 if [ "$(uname)" == "Darwin" ]; then
     MACOSX_VERSION_MIN=10.7
     CXXFLAGS="-mmacosx-version-min=${MACOSX_VERSION_MIN}"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
   md5: 6095876341956f65f9d35939ccea1a9f
 
 build:
-  number: 0
+  number: 1
   features:
     - vc9               # [win and py27]
     - vc10              # [win and py34]


jakirkham:
Could you please explain a bit more about why they must be statically linked? Does it actually present a problem to use a newer version `libstdc++` with `boost` as is? Examples would be very helpful to understand your use case.

wesm:
Let me break this into two patches. It's probably worth coming up with an example showing how incompatible libstdc++ can cause problems. 

The -fPIC issue is much more serious as the static binaries in this build cannot be linked into shared libraries.

jakirkham:
That sounds like a great idea, @wesm. Will gladly accept the `-fPIC` patch. We can then examine the `libstdc++` issue more closely.

recipe/build.sh
```diff
@@ -13,6 +13,9 @@ set -x -e
 INCLUDE_PATH="${PREFIX}/include"
 LIBRARY_PATH="${PREFIX}/lib"
 
+# Always build PIC code for easier static linking
+CXXFLAGS="${CXXFLAGS} -O3 -fPIC"
```
ccordoba12: Is `O3` safe? I think I haven't seen optimization options added to any conda package. Or is it needed by `fPIC`?

jakirkham: It shouldn't be needed by `-fPIC`.

That is a good question about `-O3`. I don't know how well that works with Boost. Have seen other libraries that have issues with this level of optimization. We should investigate this further.

wesm: I know that Impala is compiling all of its boost with -O3 and it does not cause problems, but bears more examination

jakirkham: Don't get me wrong. If it works reliably, we definitely should go for that speed up. Just want to be sure about the reliability part.

wesm:
Separated up

wesm:
@jakirkham I've been digging around for some consistent information about libstdc++; it sounds like systems with gcc 4.6 or lower as their default toolchain, you need to consistently do `-static-libstdc++` to get binaries to work. This also includes RHEL / CentOS 6. I'm going to install a CentOS 6 VM so I can see for myself on the conda forge binaries

jakirkham:
Ok, well let's go ahead with this `-fPIC` change first so that you have a better package to play with. 😄

PS - Could you please bump the build number?

wesm:
Done, and rebased

jakirkham:
LGTM

wesm:
Someone want to pull the trigger on this? @msarahan @jakirkham?

msarahan:
Thanks @wesm and @jakirkham for review.

wesm:
Thanks

PR #24, opened 2017-01-03
Use toolchain, vc packages in meta.yaml and use conda env variables

isuruf:


isuruf:
@jschueller, tests pass now.

jschueller:
thanks, it's very nice to unify the sh script between linux/osx

PR #26, opened 2017-01-04
rerender for python v3.6

scopatz:


isuruf:
Needs the VC features for `py36`

scopatz:
how do I set that? is there any documentation you can point me to?

isuruf:
I updated this wiki page for py36, https://github.com/conda-forge/staged-recipes/wiki/VC-features

ocefpaf:
Not necessarily, you can use py>=35 ;-)

isuruf:
@ocefpaf, can you edit the wiki then?

scopatz:
Better yet, this really should be in the docs on building on windows :)

scopatz:
Thanks @isuruf @ocefpaf this is now building on all platforms

jschueller:
there's no need to bump the release number as it does not change the current binaries, it just adds new builds

scopatz:
I feel safer having a new build number since it will overwrite the old packages if it doesn't

PR #31, opened 2017-01-23
Remove INCLUDE_PATH 

### LINKED ISSUE #29: Boost headers issues
It seems our recent builds of Boost now do not have headers for Boost.Python on OS X or Linux. However, on Windows it appears all of the headers are included.

cc @isuruf @jschueller

isuruf:
Fixes #29

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -19,8 +19,8 @@ if errorlevel 1 exit 1
 python -c "import os; print('_'.join(os.environ['PKG_VERSION'].split('.')[:2]))" > temp.txt
 set /p MAJ_MIN_VER=<temp.txt
 
-:: Remove non version-specific boost headers
-rm %LIBRARY_INC%\boost-%MAJ_MIN_VER%\boost
+:: Install fix-up for a non version-specific boost include
+move %LIBRARY_INC%\boost-%MAJ_MIN_VER%\boost %LIBRARY_INC%
 
 :: Move dll's to LIBRARY_BIN
 move %LIBRARY_LIB%\*vc%VS_MAJOR%0-mt-%MAJ_MIN_VER%.dll "%LIBRARY_BIN%"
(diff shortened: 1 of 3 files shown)


jakirkham:
Thanks @isuruf.

Also think we need to change this [line]( https://github.com/conda-forge/boost-feedstock/blob/4422b9d3065e32a1e377d6a283977c3183439484/recipe/bld.bat#L23 ) to use `rmdir /s /q` instead of `rm`.

jakirkham:
Think it needs a build number bump. Otherwise it looks good.

recipe/bld.bat
```diff
@@ -19,8 +19,9 @@ if errorlevel 1 exit 1
 python -c "import os; print('_'.join(os.environ['PKG_VERSION'].split('.')[:2]))" > temp.txt
 set /p MAJ_MIN_VER=<temp.txt
 
-:: Remove non version-specific boost headers
-rm %LIBRARY_INC%\boost-%MAJ_MIN_VER%\boost
+:: Install fix-up for a non version-specific boost include
+xcopy /E/Y %LIBRARY_INC%\boost-%MAJ_MIN_VER%\boost\* %LIBRARY_INC%\boost\
+rmdir /s /q %LIBRARY_INC%\boost-%MAJ_MIN_VER%\boost
```
isuruf: @jakirkham, previous approach of using `move` gave an access denied error. Using this as an alternative. If you have any other suggestions, let me know

isuruf:
Tests pass except for 2 cases failing because of corrupt downloads.

jakirkham:
Looks great @isuruf. Thanks!

recipe/bld.bat
jakirkham: It may be that there was a conflict with the existing `boost` directory from `boost-cpp`. I suppose we could have deleted it and then done this move, but your approach works just as well.

PR #34, opened 2017-03-07
Fix stderr capture

pavoljuhas:
- redirect stderr from the build commands, not the one from tee.
- let the build fail when build command `cmd` in `cmd | tee` fails.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -9,6 +9,7 @@
 # http://stackoverflow.com/questions/20108407/how-do-i-compile-boost-for-os-x-64b-platforms-with-stdlibc
 
 set -x -e
+set -o pipefail
 
 LIBRARY_PATH="${PREFIX}/lib"
 
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -28,7 +28,7 @@ LINKFLAGS="${LINKFLAGS} -L${LIBRARY_PATH}"
     --with-icu="${PREFIX}" \
     --with-python="${PYTHON}" \
     --with-python-root="${PREFIX} : ${PREFIX}/include/python${PY_VER}m ${PREFIX}/include/python${PY_VER}" \
-    | tee bootstrap.log 2>&1
+    2>&1 | tee bootstrap.log
 
 ./b2 -q \
     variant=release \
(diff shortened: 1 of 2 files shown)


scopatz:
This seems to have some issues.

pavoljuhas:
ci/circleci and appveyor failed to get boost sources from sourceforge, which was likely down at that time.  travis-ci jobs are still in their Mac OS X queue.  I don't see anything related to the PR.
Or do you have something else in mind besides these CI failures?

scopatz:
I didn't look at why they failed.  Just wanted to make sure that you knew that they did.

isuruf:
circleci error is not a download failure. it seems with your changes, there is no output for more than 10 minutes and circleci will terminate any job with no output for 10 minutes.

pavoljuhas:
The last output in the circleci log is
```
...
Downloading http://sourceforge.net/projects/boost/files/boost/1.63.0/boost_1_63_0.tar.bz2
```
I read it that the download got either completely stuck or was taking too long.  Right now I am getting 6 minute download time from that URL.  This is a problem of sourceforge servers, because a similar-sized file on GitHub downloads in 30 seconds.  It is quite likely the download from SF took over 10 minutes yesterday.  

Can you or @scopatz restart the circleci to see if it would get past the download step now?
If not I can do so myself by pushing an empty commit.

CJ-Wright:
@pavoljuhas You can just cycle the PR (close and reopen) to restart the CIs.

isuruf:
@pavoljuhas, you are right. I changed the url and it works. @conda-forge/boost is it okay if I change the url like this?

jakirkham:
Sure. TBH SourceForge download failures with Boost have been a real drag.

Though if we are planning on making this a long term solution, maybe we should have a more principled naming scheme with the releases. What do you think?

isuruf:
I had release called `sources` and we can attach any of the boost sources we need to the same release. Thoughts?

pavoljuhas:
We might want to check with [boostorg/release-tools](https://github.com/boostorg/release-tools) maintainers.

@grafikrobot, @Beman - is there any official mirror of boost releases that is faster and more reliable than sourceforge?  The CI builds at conda-forge have frequent break downs due to failed downloads from sourceforge.

pavoljuhas:
Should I rebase this to drop the dummy commit f753c67?

jakirkham:
Sure.

pavoljuhas:
Done.

PR #39, opened 2017-05-04
Write user-config with python settings

jschueller:


isuruf:
Does this require a build number bump?

jschueller:
I dont know

jschueller:
@isuruf, I bumped the build number as now there's no more libboost_python.lib for python3, only libboost_python3.lib, which is consistent with the unix builds.

PR #40, opened 2017-06-02
Add patch for boost::serialization

SoapZA:
* Fixes missing include:
  https://svn.boost.org/trac/boost/ticket/12516


SoapZA:
@jschueller do I also need to send this fix in for boost-cpp?

jakirkham:
Actually it needs to be moved over to boost-cpp. This is basically just the Python components of Boost with the rest pulled in as dependencies. So would recommend closing this out and adding a patch to boost-cpp.

jschueller:
Also you could mention the commit and ticket in the meta.yaml next to the patch (in your next PR :)

PR #43, opened 2017-06-10
Patch BOOST_LIB_NAME for better Python 3 support

jakirkham:
Applies the patch from PR ( https://github.com/boostorg/python/pull/135 ) to correct an issue where Boost incorrectly links against `boost_python` libraries on Python 3 instead of `boost_python3` libraries. More info can be found in this related issue ( https://github.com/boostorg/python/issues/127 ).

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -17,7 +17,7 @@ source:
     - PR_135.diff
 
 build:
-  number: 1
+  number: 2
   features:
     - vc9               # [win and py27]
     - vc14              # [win and py>=35]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -9,6 +9,12 @@ source:
   fn:  {{ filename }}
   url: https://dl.bintray.com/boostorg/release/{{ version }}/source/{{ filename }}
   sha256: 7bcc5caace97baa948931d712ea5f37038dbb1c5d89b43ad4def4ed7cb683332
+  patches:
+    # Applies a patch to correct `BOOST_LIB_NAME` on Python 3.
+    #
+    # xref: https://github.com/boostorg/python/pull/135
+    #
+    - PR_135.diff
 
 build:
   number: 1
--- a/recipe/PR_135.diff
+++ b/recipe/PR_135.diff
@@ -1,0 +1,17 @@
+diff --git include/boost/python/detail/config.hpp include/boost/python/detail/config.hpp
+index c92ecb32b..3e4b7c9e5 100644
+--- include/boost/python/detail/config.hpp
++++ include/boost/python/detail/config.hpp
+@@ -105,7 +105,11 @@
+ // Set the name of our library, this will get undef'ed by auto_link.hpp
+ // once it's done with it:
+ //
+-#define BOOST_LIB_NAME boost_python
++#if PY_MAJOR_VERSION == 2
++#  define BOOST_LIB_NAME boost_python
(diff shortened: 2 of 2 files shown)


recipe/meta.yaml
```diff
@@ -9,9 +9,15 @@ source:
   fn:  {{ filename }}
   url: https://dl.bintray.com/boostorg/release/{{ version }}/source/{{ filename }}
   sha256: 7bcc5caace97baa948931d712ea5f37038dbb1c5d89b43ad4def4ed7cb683332
+  patches:
+    # Applies a patch to correct `BOOST_LIB_NAME` on Python 3.
+    #
+    # xref: https://github.com/boostorg/python/pull/135
```
jakirkham: xref: https://github.com/boostorg/python/pull/135

recipe/meta.yaml
```diff
@@ -9,9 +9,15 @@ source:
...
+    #
+    - PR_135.patch
```
jakirkham: xref: https://github.com/boostorg/python/pull/135

recipe/meta.yaml
```diff
@@ -9,9 +9,15 @@ source:
   fn:  {{ filename }}
   url: https://dl.bintray.com/boostorg/release/{{ version }}/source/{{ filename }}
   sha256: 7bcc5caace97baa948931d712ea5f37038dbb1c5d89b43ad4def4ed7cb683332
+  patches:
+    # Applies a patch to correct `BOOST_LIB_NAME` on Python 3.
+    #
+    # xref: https://github.com/boostorg/python/pull/135
+    #
+    - PR_135.diff
```
jakirkham: xref: https://github.com/boostorg/python/pull/135

jakirkham: This has been accepted upstream.

jakirkham:
Could you please take a look @jschueller?

isuruf:
Looks good to me. FWIW, I think we should just create the symlink libboost_python -> libboost_python3 to make it easier to package software depending on this. We'll have to patch software that haven't had a release after boost 1.64 release.

jakirkham:
Thanks for taking a look @isuruf. I wouldn't have been opposed to it, but this should hopefully make that unnecessary.

jschueller:
+1, now the scary part: pin 1.64 in all feedstock :)

jakirkham:
Thanks for the reviews.

Yeah we need to get that pinning script back up in running. If either of you have time to look at stuff related to the pinning script (and maybe nudge it along), that would be really helpful. Happy to provide pointers to relevant material.

jschueller:
Is it some sort of bot ?

jakirkham:
Yeah, there is a pinning bot. At the core of it, it runs this [script]( https://github.com/conda-forge/conda-forge.github.io/blob/21bdc7b1b6ef981f1ebcebb7307bd2bb5c6a7cf2/scripts/pin_the_slow_way.py ). Unfortunately it has gone basically dormant because of a bug fix that created a new bug. ( https://github.com/conda-forge/conda-forge.github.io/pull/256 ) Namely it only pins unpinned dependencies, but is unable to update existing pinning dependencies. It basically comes down to [one bad line]( https://github.com/conda-forge/conda-forge.github.io/blob/21bdc7b1b6ef981f1ebcebb7307bd2bb5c6a7cf2/scripts/pin_the_slow_way.py#L293 ) of regex.

PR #44, opened 2017-07-19
Sane numpy matrix

ocefpaf:
@jjhelmus I'm stealing your idea here :wink: 

I rebuilt everything from our internal system and it worked wonders! No reason not to use it in `conda-forge` too.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -35,9 +35,9 @@ requirements:
 
   run:
     - python
-    - numpy 1.7.*  # [py27]
-    - numpy 1.9.*  # [py35]
-    - numpy 1.11.*  # [py36]
+    - numpy >=1.7  # [py27]
+    - numpy >=1.9  # [py35]
+    - numpy >=1.11  # [py36]
     - vc 9              # [win and py27]
     - vc 14             # [win and py>=35]
     - boost-cpp {{ version }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -17,7 +17,7 @@ source:
     - PR_135.diff
 
 build:
-  number: 2
+  number: 3
   features:
     - vc9               # [win and py27]
     - vc14              # [win and py>=35]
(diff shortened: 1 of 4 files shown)


recipe/meta.yaml
```diff
@@ -26,14 +26,18 @@ requirements:
...
     - vc 9              # [win and py27]
     - vc 14             # [win and py>=35]
     - boost-cpp {{ version }}
 
   run:
     - python
-    - numpy x.x
+    - numpy 1.7.*  # [py27]
+    - numpy 1.9.*  # [py35]
+    - numpy 1.11.*  # [py36]
```
jakirkham: Aren't these suppose to be lower bounded not pinned to a major minor version? 😕

xref: https://github.com/conda-forge/pandas-feedstock/pull/32/files#diff-e178b687b10a71a3348107ae3154e44cR27

ocefpaf: Only when the PR is ready. Not when I want to ensure that the same `numpy` is used to check the linkages and the `mkl` vs `openblas` situation.

jakirkham: I don't think Boost is using a BLAS, but am all for being cautious. Thanks for checking. :)

recipe/meta.yaml
```diff
@@ -26,14 +26,18 @@ requirements:
...
     - vc 9              # [win and py27]
     - vc 14             # [win and py>=35]
     - boost-cpp {{ version }}
 
   run:
     - python
-    - numpy x.x
+    - numpy >=1.7  # [py27]
+    - numpy >=1.9  # [py35]
+    - numpy >=1.11  # [py36]
```
ocefpaf: @jjhelmus I got `numpy 1.7.*` working with `openblas` for `py27` but not `py35`. I wonder if we should try `numpy 1.9` everywhere to simplify this. What do you think?

isuruf:
@ocefpaf, can you open a PR for the docs repo?

ocefpaf:
> @ocefpaf, can you open a PR for the docs repo?

Should've done that first :wink: 

Working on it now. I'll ping you for a review soon.

PR #54, opened 2018-04-18
boost v1.67.0

tadeu:
Based on PR #52

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ os: osx
 osx_image: xcode6.4
 
 env:
+  matrix:
+    
+    - CONDA_PY=27
+    - CONDA_PY=35
+    - CONDA_PY=36
   global:
     # The BINSTAR_TOKEN secure variable. This is defined canonically in conda-forge.yml.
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -7,11 +7,6 @@ os: osx
 osx_image: xcode6.4
 
 env:
-  matrix:
-    
-    - CONDA_PY=27
-    - CONDA_PY=35
-    - CONDA_PY=36
```
jakirkham: Hmm...do we know why these disappeared?

appveyor.yml
```diff
@@ -17,22 +17,6 @@ environment:
...
-      CONDA_PY: 35
-      CONDA_INSTALL_LOCN: C:\\Miniconda35-x64
-
-    - TARGET_ARCH: x86
-      CONDA_PY: 36
-      CONDA_INSTALL_LOCN: C:\\Miniconda36
-
-    - TARGET_ARCH: x64
-      CONDA_PY: 36
-      CONDA_INSTALL_LOCN: C:\\Miniconda36-x64
```
jakirkham: Same story here. :/

jakirkham:
@conda-forge-admin, please rerender.

jakirkham:
Thanks for picking this up, @tadeu.

Something is off with the re-rendering. So asked the bot for a new one. Let's see how that goes.

.travis.yml
tadeu: Good catch, I didn't check these other changes, I've just continued from where the bot PR stopped.

jakirkham: Yeah, have seen these issues in the past with re-rendering. Though that is not a common occurrence with the re-rendering bot or manual re-renderings any more. Not sure why the autoupdate bot is running into this issue, but have seen it in a couple of cases. Normally our bot or manually re-rendering has fixed it afterwards. Could be anything from a connectivity issue or suppressed error.

Letting you know for awareness, @CJ-Wright. Probably not worth worrying about to much, but if there is an issue to track this that would be good for now. If you have any ideas, they would of course be welcome, but just being aware is probably good for now given cb3 and cs3 are rolling out too.

isuruf: it was because `boost-cpp` PR was not merged at the time.

jakirkham:
Seems the re-rendering bot fixed it. So no need to worry about it. 😌

.travis.yml
jakirkham: Good point. Thanks for mentioning that.

tadeu:
Is there something else needed for the merge?

jakirkham:
Thanks @tadeu. :)

PR #55, opened 2018-05-18
Added  -std=c++14 flag

xmnlab:


scopatz:
I think we need to wait for the new compilers

xmnlab:
@scopatz 

is there some idea about when the new compilers will be available?

scopatz:
Folks are working on it now, so "soon" hopefully

jschueller:
when using the new compiler syntax, it pull gcc 7.2 which default to using c++14, so this should be useless now, reopen if needed

xmnlab:
Thanks for the update!

PR #131, opened 2022-01-26
Backported patch to fix double-digit minor python version in BoostConfig.cmake

ckoern:
Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

Boost 1.74 fails to recognise python 3.10, which was fixed in 1.78. As this bug makes the build of other feedstocks fail for python 3.10 (e.g. conda-forge/geant4-feedstock#33), it should be worth backporting the fix

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -11,43 +11,43 @@ jobs:
       linux_64_numpy1.18python3.7.____cpython:
         CONFIG: linux_64_numpy1.18python3.7.____cpython
         UPLOAD_PACKAGES: 'True'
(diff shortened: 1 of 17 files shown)


ckoern:
@conda-forge-admin, please rerender

xhochy:
This should also be backported to 1.76 probably?

chrisburr (approved):

ckoern:
I have realised, that the BoostConfig.cmake is not packed in the final package output but that this file is instead provided by `boost-cpp`, so this PR actually has no effect. Sorry about that!

I will make a PR in `boost-cpp` instead.

PR #154, opened 2022-10-27
Rebuild pypy38 0 1 h3f9770

xhochy:
Fixes #152


jakirkham:
Thanks Uwe! 🙏 

This seems reasonable. Do we want to make this change to `main` as well?

Issue #159, opened 2022-12-22
Missing static libraries?


gregdenay:
### Comment:

I am trying to update a bioconda recipe for a package requiring the BOOST static libraries to build (SKESA).
However, it seems that the static libraries are missing from the /lib folder of the conda environment.

On debian 10:
```
 mamba create -n boost boost zlib
```

~/anaconda3/envs/boost/include/boost/ exists
~/anaconda3/envs/boost/lib contains the libboost_xxx.so files but no libboost_xxx.a file.

Looking at the recipe here it seems like the static libraries should be installed.  I am a bit confused with the otrher stock (boost-cpp) for which the static libraries are not installed. The mamba log lists both packages:

```
                  __    __    __    __
                 /  \  /  \  /  \  /  \
                /    \/    \/    \/    \
███████████████/  /██/  /██/  /██/  /████████████████████████
              /  / \   / \   / \   / \  \____
             /  /   \_/   \_/   \_/   \    o \__,
            / _/                       \_____/  `
            |/
        ███╗   ███╗ █████╗ ███╗   ███╗██████╗  █████╗
        ████╗ ████║██╔══██╗████╗ ████║██╔══██╗██╔══██╗
        ██╔████╔██║███████║██╔████╔██║██████╔╝███████║
        ██║╚██╔╝██║██╔══██║██║╚██╔╝██║██╔══██╗██╔══██║
        ██║ ╚═╝ ██║██║  ██║██║ ╚═╝ ██║██████╔╝██║  ██║
        ╚═╝     ╚═╝╚═╝  ╚═╝╚═╝     ╚═╝╚═════╝ ╚═╝  ╚═╝

        mamba (1.0.0) supported by @QuantStack

        GitHub:  https://github.com/mamba-org/mamba
        Twitter: https://twitter.com/QuantStack

█████████████████████████████████████████████████████████████


Looking for: ['boost', 'zlib']

pkgs/main/linux-64                                            No change
bioconda/linux-64                                             No change
bioconda/noarch                                               No change
r/linux-64                                                    No change
pkgs/main/noarch                                              No change
pkgs/r/linux-64                                               No change
pkgs/r/noarch                                                 No change
r/noarch                                                      No change
conda-forge/noarch                                  10.7MB @   2.8MB/s  4.1s
conda-forge/linux-64                                28.7MB @   5.1MB/s  6.6s
Transaction

  Prefix: /home/debian/anaconda3/envs/boost

  Updating specs:

   - boost
[...]

Issue #119, opened 2021-10-11
Missing CMake Config for Static Variant


nmaxwell:
- [ x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue:

(see full environment list below)

This error occurs while compiling either osmium tool (https://github.com/osmcode/osmium-tool) or libLAS (https://github.com/libLAS/libLAS).
Both libraries use CMake, and set "Boost_USE_STATIC_LIBS=ON". All fine and good, I've verified that this 'boost' package from conda-forge provides both static and shared libraries. The problem is caused by missing CMake configuration from Boost, for the shared variant. 

For example, for boost-chrono, in `D:\cvrs-root-test\Library\lib\cmake\boost_chrono-1.74.0\boost_chrono-config.cmake`, there is this glob logic to find all build variants: 
```file(GLOB __boost_variants "${CMAKE_CURRENT_LIST_DIR}/libboost_chrono-variant*.cmake")```

 But this is the directory listing under `Library\lib\cmake\boost_chrono-1.74.0`

![boost](https://user-images.githubusercontent.com/132289/136825628-ad260738-4253-4396-be6d-18f72912d918.PNG)

So it has `libboost_chrono-variant-shared.cmake`, but seems to be missing something like `libboost_chrono-variant-static.cmake`. This causes the `find_package(Boost...` call to fail when Boost_USE_STATIC_LIBS is ON. 

I've tried building with it off, but get other linking errors. I have verified that I am able to compile and link some simple boost example programs in the environment (with dynamic linking). So I think that means these libraries really do need static linking, I'm not exactly clear on why though.

Full error:
```
CMake Error at D:/cvrs-root-test/Library/lib/cmake/Boost-1.74.0/BoostConfig.cmake:141 (find_package):
  Found package configuration file:

    D:/cvrs-root-test/Library/lib/cmake/boost_program_options-1.74.0/boost_program_options-config.cmake

  but it set boost_program_options_FOUND to FALSE so package
  "boost_program_options" is considered to be NOT FOUND.  Reason given by
  package:

  No suitable build variant has been found.

  The following variants have been tried and rejected:

  * boost_program_options.lib (shared, Boost_USE_STATIC_LIBS=ON)
```

```
$ conda info

     active environment : D:\cvrs-root-test
    active env location : D:\cvrs-root-test
            shell level : 2
[...]

PR #164, opened 2023-05-03
Boost unification

### LINKED ISSUE #53: Consolidate with boost-cpp
The AnacondaRecipes version of this feedstock (https://github.com/anacondarecipes/boost-cpp-feedstock) incorporates boost-python and libboost, as well as "boost" and "boost-cpp" as metapackages.  As part of the unification with defaults, we should discuss how to deprecate this repo and have only one boost feedstock that has the package splitting.

h-vetinari:
Implementation for https://github.com/conda-forge/boost-cpp-feedstock/issues/137

Closes #53
Closes https://github.com/conda-forge/boost-cpp-feedstock/issues/25

As discussed in that issue, this PR:
* unifies the feedstocks for [boost](https://github.com/conda-forge/boost-feedstock) & [boost-cpp](https://github.com/conda-forge/boost-cpp-feedstock) in this feedstock (due to better name)
* renames `boost-cpp` -> `libboost` & `boost` -> `liboost-python` (maintaining the old output names as wrappers for now)
* adds a run-export to the new `libboost` (while `boost-cpp` stays unchanged without one)
* introduces a header-only version `libboost-headers` that will fill the role of the run-export-less `boost-cpp`

The history of https://github.com/conda-forge/boost-cpp-feedstock is preserved through a [merge](https://github.com/conda-forge/boost-feedstock/pull/164/commits/ca2d109fbf5b0fe41bdfa10eb96d8169ea6d4517) (after some preparatory commits to reduce collisions). This makes the PR look unwieldy but the changes before the merge aren't very interesting. The upside is that this allows digging into the full archeology in one place (with the only snag that GH now resolves the PR links incorrectly).

Implementation notes:
* `libboost` depends on `libboost-headers`, so the headers are really only packaged once
* I tried but quickly gave up to build all three outputs at once; I believe this would be hard both because boost's `b2` built tool doesn't seem to support only-build-don't-install so well, and more importantly, because we need to build `libboost-python` _per python version_.
* Therefore, `libboost` & `libboost-headers` are built only once in `build.sh` / `bld.bat` (essentially the build scripts from boost-cpp), with the only substantial change that they're now "installed" into a temporary prefix from which we can then actually install the respective files into our $PREFIX per output
* `libboost-python` essentially keeps the build scripts of boost, only that we now reuse the `b2` that was bootstrapped during the global build phase (this now allows full cross-compilation; previously ppc was emulated here).
* There's still substantial duplication between the (configuration parts of) `build-py.{sh,bat}` and the global build scripts, but for now I've tried to keep the changes to the respective histories to a minimum.
* OTOH, the tests should be cleaned up nicely now.
* Switched `libboost-python` to only package dynamic libs, and add a run-export. Also, while it previously depended on the full `boost-cpp`, the link-check says the libs aren't necessary, so switch to depend on `libboost-headers` instead.

PTAL @conda-forge/boost @conda-forge/boost-cpp
CC @conda-forge/core


beckermr:
I want to make sure I understand one item. The old boost-cpp has a constraint that libboost<0. The new boost-cpp has an exact pin. So the run constraint on the conda-forge libboost of boost-cpp at the same version will force loading the new boost-cpp. Am I following this correctly?

We may want to put a note in the recipe to always keep the exact pin. Otherwise I suppose we could pull in libboost from defaults in some conda configurations (though this would not be officially supported).

h-vetinari:
> The old boost-cpp has a constraint that libboost<0. The new boost-cpp has an exact pin.

Yes, otherwise it could not be a wrapper around `libboost`.

> So the run constraint on the conda-forge libboost of boost-cpp at the same version will force loading the new boost-cpp. Am I following this correctly?

Yes, the constraint ensures that we can only co-install `libboost` & `boost-cpp` when both have equal version, and since our `libboost` that carries this constraint only exists since this PR (and `boost-cpp` wraps _that_), it remains impossible to co-install either one with old `libboost` from Anaconda.

The same sort of constraint is already working for similar (though less ambitious) renamings of abseil, arrow & grpc.

beckermr:
I see. I see the current comment but it doesn't explain in detail. It may be worth putting that explanation in a comment. 

Thanks a bunch for the work here. Lgtm!

recipe/build.sh
```diff
@@ -67,13 +67,15 @@ fi
...
     include="${INCLUDE_PATH}" \
     cxxflags="${CXXFLAGS}" \
     linkflags="${LINKFLAGS}" \
     --layout=system \
-    --with-python \
     -j"${CPU_COUNT}" \
-    install 2>&1 | tee b2.log
+    install
+
+# python headers packaged in build-py.sh (because we need to build per python version)
```
isuruf, line 79: Are you sure about this?

h-vetinari: Perhaps "need" is the wrong word. Historically, this has been packaged per python version, which also ends up reflected in the library name (e.g. `libboost_python39.so`). I kept that setup, since I wanted to minimise changes while merging the feedstocks, and because I don't know if it would actually be possible to do just once (which would be nice of course!)

h-vetinari:
I'm taking the hearts as indications that things are not looking too bad. ;-)

(@isuruf, of course I'm willing to address your feedback if you think something needs to be done; though I think leaving the `-python` builds more or less as-is should be OK for a first step)

Opened PR for the remaining pre-requisite: https://github.com/conda-forge/feedstock-outputs/pull/47

isuruf:
Let's give this some time. There's no particular hurry for this and I'll give some feedback later.

h-vetinari:
> Let's give this some time. There's no particular hurry for this and I'll give some feedback later.

In terms of timeline, I'd like to get the migrations for boost (and flang!) done before we have the Python 3.12 migration in ~October, to avoid having to deal with these big migrations getting mixed, which caused some [heartache](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3268#issuecomment-1352077551) last time. Also we're kinda due a boost migration anyway, as we're on 1.78 now and 1.82 is [ready](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4358) (last [increment](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2757) was +4 in minor number as well after a bunch of discussion).

recipe/meta.yaml
```diff
@@ -11,45 +21,226 @@ source:
...
+        - numpy                                  # [build_platform != target_platform]
+        - {{ compiler('cxx') }}
+      host:
+        - python
+        - numpy
+        - {{ pin_subpackage("libboost-headers", exact=True) }}
+      run:
+        - python
+        - {{ pin_compatible('numpy') }}
+        - {{ pin_subpackage("libboost-headers", exact=True) }}
```
sdebionne, line 193: Rephrasing https://github.com/conda-forge/boost-cpp-feedstock/issues/137#issuecomment-1539573207

```suggestion
```

h-vetinari: I'm not against loosening the pins on the headers in principle, but I don't know boost well enough to know that they'll always be compatible. For context there was a [response](https://github.com/conda-forge/boost-cpp-feedstock/issues/137#issuecomment-1540009887) on that in the other thread:
> @beckermr: I'd prefer exact pins for now with the idea that we can relax them later if it is working.

-------

> @sdebionne: The need for `devel` packages, not specific to Boost, might be out of scope of this MR though.

Yeah, I think this PR is already very large and so far has the scope as defined in https://github.com/conda-forge/boost-cpp-feedstock/issues/137: merge feedstocks, rename packages, add run-exports.

I'll do my best to ensure it does these things to everyone's satisfaction, but I don't want to overload this with refactoring how `libboost-python` works, adding `-devel` packages (another discussion point for consistency / bikeshedding), etc. That should be left to follow-ups.

sdebionne: > I'm not against loosening the pins on the headers in principle

Actually, I suggested to remove the requirement alltogether (and only have it for the `-devel` packages).

> I don't know boost well enough to know that they'll always be compatible

The headers _must_ match the version of the library.

> adding -devel packages (another discussion point for consistency / bikeshedding)

Ok, let's discuss this in a new issue.

recipe/bld.bat
```diff
@@ -1,31 +1,39 @@
...
 :: Build step
 .\b2 install ^
     --build-dir=buildboost ^
-    --prefix=%LIBRARY_PREFIX% ^
+    --prefix=temp_prefix ^
     toolset=msvc-%VS_MAJOR%.0 ^
     address-model=%ARCH% ^
     variant=release ^
     threading=multi ^
-    link=static,shared ^
```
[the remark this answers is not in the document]
beckermr: Our policy is to put them in a separate package.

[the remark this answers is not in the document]
sdebionne: I guess you meant subpackage? Then both variant (link=static,shared) needs to be compiled and new `output`s defined for the static libraries?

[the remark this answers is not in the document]
beckermr: Yes.

recipe/build.sh
isuruf: Let's keep all the headers at the same place. We don't have to keep carrying tech debt forever.

h-vetinari: > We don't have to keep carrying tech debt forever.

I don't have all the background on this, but I presume the headers for `libboost-python` are not actually dependent on the python version, only the compiled libraries. So - IIUC - we still have to compile `libboost-python` per python version, but we can ensure that the python headers are already present in `libboost-headers`.

That sounds like a clear improvement indeed, and happy to give it a shot.

recipe/meta.yaml
isuruf: > Yeah, I think this PR is already very large and so far has the scope as defined in https://github.com/conda-forge/boost-cpp-feedstock/issues/137: merge feedstocks, rename packages, add run-exports and an output for only the headers.

I'm against this line of thinking. As I said earlier, there's no need to keep technical debt forever. I agree with @sdebionne's suggestion here.

h-vetinari: > As I said earlier, there's no need to keep technical debt forever.

There's also no need to require one PR to clean up all extant technical debt. In fact, that's not a reasonable demand _at all_. I'm keeping the commit history sane, but the more this (or any other large) PR does at once, the harder it will be to reason about if anything changes or breaks.

Standard development practices require PRs to be focused on one thing for a reason.

I've taken on your request of where to place the headers for the python libs, as I think it's broadly in scope of this PR. But changing what `libboost-python` pins or brings along is another question. You cannot compile against it without the headers. So if it's then about having a lighter run-dependence, that's a separate improvement.

Because deferring the discussion of how to pin the headers vs. `libboost-python` is in no way "keeping technical debt forever". It's just one PR away after this one gets merged and doesn't blow up anything. Plus I agree with @beckermr that it's much easier to loosen requirements afterwards.

h-vetinari: > Standard development practices require PRs to be focused on one thing for a reason.

For a bit more context: I recognize of course that in contrast to any upstream repo merging a PR, merging this one will publish artefacts. However, the PR is aimed as much as possible at not breaking anything, just changing behind the scenes through which output a given file gets installed.

In other words, recipes or other consumers of `boost-cpp` should get the exact same files as before[^1]. For `boost`, we've now removed the non-python libraries, but if that's a concern I'm happy to revert this (765cc2cf9d24373bb990ebc82d04b99ebee0d0d5).

[^1]: plus the headers for boost python, after implementing your request for exactly that

This PR aims to achieve a milestone, not the end goal. And shuffling files between outputs with very low risk of breakage (especially as 1.82 is only 2 weeks old and not migrated yet) is IMO a very good way point to achieve before embarking on further improvements. That's why I'm so reluctant to layer more changes on top here that aren't absolutely necessary for the [goals](https://github.com/conda-forge/boost-cpp-feedstock/issues/137) that were agreed upon. Not because I'm lazy, but because I think it's detrimental to maintenance later as well as actually pulling off this large tech debt clean-up.

But if this becomes the stumbling block, then perhaps we publish these builds to a `boost_dev` label, so that we can test them in selected feedstocks and iterate as necessary (but with intermediate way points & easily testable artefacts).

In any case, it would be beneficial to write up a plan of action of what else you think should be done (whether compatibly or incompatibly with the current boost setup; as I keep saying this PR is aiming at the former), rather than saying "this doesn't address all of our technical debt" in this PR - I hope we can agree it's a step in the right direction at least, and that we can build on that.

> I've taken on your request of where to place the headers for the python libs, as I think it's broadly in scope of this PR. 

This is passing now.

h-vetinari:
> In terms of timeline, I'd like to get the migrations for boost (and flang!) done before we have the Python 3.12 migration in ~October, to avoid having to deal with these big migrations getting mixed

Python 3.12b1 got [released](https://github.com/python/cpython/releases/tag/v3.12.0b1) today. A migration is still a while away, but it'd be nice to leave a bit of buffer. Can we start wrapping up the review here?

isuruf:
Let's merge this with a different label than `main` and work on it before releasing it into `main`.

h-vetinari:
> Let's merge this with a different label than `main` and work on it before releasing it into `main`.

OK. Could you write up a list which work is still necessary in your view? It should already be as capable as the current boost setup, though I guess we could add static outputs? I'm willing to do it, but I'd need to know what you think should be done.

h-vetinari:
Note that this still requires https://github.com/conda-forge/feedstock-outputs/pull/47 before merging.

isuruf:
How about the following?

- boost-headers (only headers)
- libboost (only shared libraries)
- libboost-python (only shared libraries)
- libboost-devel (boost-headers + libboost with run_exports + pkg-config & cmake files)
- libboost-python-devel (boost-headers + libboost-python with run_exports  + pkg-config & cmake files)

h-vetinari:
> How about the following?

I'm fine with that separation, though I'd have a preference to keep `libboost-headers` (rather than `boost-headers`), for consistency with the other `libboost-*` outputs.

The one thing that I'm not thrilled about is that it opens another can of worms, which is about naming consistency between the different variants `foo-devel`, `foo-dev`, `foodev`, etc. for spelling this concept. The former is very widespread in distros of course, so it seems like a natural choice, but I'd like us to decide on one way to do these, rather than have a hodgepodge of inconsistent namings and ways in which the files are sliced into outputs.

I also noticed that there's a deferred [CFEP](https://github.com/conda-forge/cfep/blob/main/cfep-20.md) on this topic by @SylvainCorlay; I haven't read all the discussion in https://github.com/conda-forge/cfep/pull/39 in detail, but it looks like another big topic, and I doubt we'd come to a solution quickly (e.g. the new CUDA setup [introduced](https://github.com/conda-forge/cuda-libraries-dev-feedstock) [two](https://github.com/conda-forge/cuda-nvml-dev-feedstock) `-dev` feedstocks recently, we have `llvmdev` and `clangdev` that sit really low in our stack, etc.)

In short, I don't see why we _have_ to attach this to the unification here. The same approach (e.g. refactor and migrate with renaming) could be used for a future boost-version when that discussion has taken place.

With your proposal we're IMO risking to solidify that divergence around `-devel` namings, or stalling this effort for months or longer. If people only want the renaming hassle once (or insist to do this concurrently for other reasons), I can live with moving forward with this PR+`-devel`, though my preference would be to keep it separate.

h-vetinari:
Another point in the `-devel` discussion is that distros usually also package the static libraries (boost in ubuntu is a sprawling set of outputs, but here's an [example](https://packages.ubuntu.com/jammy/amd64/libboost-numpy1.74-dev/filelist); here's [another](https://packages.ubuntu.com/jammy/amd64/libre2-dev/filelist) for a simpler lib). Aside from going against CFEP-18 (so it should be something exceptional), we generally don't want to have a run-export for something that builds against the static libs. So I think it'd makes sense that feedstocks continue into opting into providing `foo-static` as a separate output (we ~25 libs that do this [already](https://github.com/search?q=org%3Aconda-forge+%2F-+name%3A.*-static%2F+path%3Arecipe%2Fmeta.yaml&type=code)), rather than absorbing static libs into `foo-devel`.

This is not because I want to argue that point here, but to illustrate that "we should have a `-devel` package" is fraught with all the other questions about this topic this would open (or alternatively, risk further balkanizing the situation). And it's a question that should be considered, as the desire for static boost has been mentioned above already (xref #119 & #159)

It's also worth noting that any future renaming from the state of this PR to @isuruf's proposal would be mechanically very simple, rather than the more involved logic ("is it also in run or just in host?") that we need for https://github.com/regro/cf-scripts/pull/1668.

h-vetinari:
From the discussion in the core call last week, the message was:
* disruption only once
* there must be a lib output without headers
* people didn't feel enthusiastic about either `-dev` or `-devel`, the option was brought up to keep `boost-cpp`.

Based on these inputs, here's an updated proposal:

| output | content | run-depends | run-exports |
|-|-|-|-|
| `libboost-headers` | headers | - | - |
| `libboost` | shared libraries | - | - |
| `libboost-python` | shared python libraries | - | - |  |
| `boost-cpp` | pkg-config & cmake files | `libboost-headers`<br/>`libboost` | `libboost` | 
| `boost-python` | pkg-config & cmake files | `libboost-headers`<br/>`libboost-python` | `libboost-python` | 

Some observations:
* Anaconda's `libboost` would then correspond to `boost-cpp` again (as previously); name would stay the same within conda-forge, but now have a run-export (so we're relying on the piggyback migrator to take care of the behaviour change)
* `libboost-python` & `boost-python` fit consistently into the naming pattern, but matches neither existing `boost` nor Anaconda's `py-boost`
* It would be possible to extend this scheme to include `libboost-static` & `libboost-python-static`, which however wouldn't/shouldn't (IMO) be part of the "devel" packages.
* I have a preference for the headers to contain the "lib" prefix, but if someone strongly prefers `boost-headers`, I'm fine with that.

Aside from the impact on _not_ achieving mostly unified naming with Anaconda (which is a somewhat foregone conclusion now that we need an extra two outputs), is this something that people could agree on?

h-vetinari:
> Aside from the impact on _not_ achieving mostly unified naming with Anaconda (which is a somewhat foregone conclusion now that we need an extra two outputs), is this something that people could agree on?

Gentle ping @isuruf @beckermr @mbargull @chenghlee @conda-forge/boost

(I've updated the table to show how the optional `-static` outputs would fit in).

h-vetinari:
Boost 1.83beta just got released, so it looks like we can just move the setup with the new version. Still, could I please ask for feedback on the approach here?

h-vetinari:
I'm running out of ways to ask for feedback here, which is quite frustrating. For those who don't want to write out a full response, please "emoji-react" to this post:

* 👍 The above [plan](https://github.com/conda-forge/boost-feedstock/pull/164#issuecomment-1618232557), _**without**_ `libboost-{python-,}static`
* ❤️ The above [plan](https://github.com/conda-forge/boost-feedstock/pull/164#issuecomment-1618232557), _**with**_ `libboost-{python-,}static`
* 🚀 Use @isuruf's [suggestion](https://github.com/conda-forge/boost-feedstock/pull/164#issuecomment-1596200934) verbatim
* 🎉 Use @isuruf's [suggestion](https://github.com/conda-forge/boost-feedstock/pull/164#issuecomment-1596200934), but with `-dev` instead of `-devel`
* 😕 Revive [CFEP-20](https://github.com/conda-forge/cfep/blob/main/cfep-20.md)
* 😄  Keep things as in the current state of this PR (no additional output without headers)
* 👀 Something else entirely

All but "😄" fail the "unify with Anaconda naming" goal, but I guess that's for Anaconda to deal with.

beckermr:
Hey! I'm sorry I've been silent on this. In general I'd say that core members consult with one another and try to reach consensus. At some point though, the person doing the work needs to be unblocked. So IMHO you should feel free to simply proceed unless there are very strong objections. If someone has a strong objection, they should be willing to do the work in a timely manner. 

I'm personally fine with any of the options and happy to help with bot work.

isuruf:
> At some point though, the person doing the work needs to be unblocked. So IMHO you should feel free to simply proceed unless there are very strong objections. If someone has a strong objection, they should be willing to do the work in a timely manner.

I highly disagree with this. That's disrespectful to the maintainers of this feedstock. If the maintainers have a strong objection, the PR author should address it. Not the other way around. Please don't set such disrespectful precedents.

h-vetinari:
> That's disrespectful to the maintainers of this feedstock. If the maintainers have a strong objection, the PR author should address it. Not the other way around. Please don't set such disrespectful precedents.

That is... really far off from what happened. 5 weeks ago this was again discussed in the core call with you present, and the plan I wrote up was based on _your_ suggestion to keep using `boost-cpp` (after I raised a concern about `-devel` vis-à-vis CFEP-20).

What I find disrespectful is raising a blocker in a PR and then not responding anymore, for weeks or months (until something moves and you're back immediately to reiterate your objection). This is a situation that I find myself with you with some regularity[^1], and I honestly don't know what I've done to deserve that.

It's disheartening from start to finish.

[^1]: I could name 10 PRs off the top of my head, but my point is not to dig up old grievances, just to underscore that I'm not responding to a single incident.

jaimergp:
I think the proposal so far has tried to accommodate every possible angle of feedback while pushing for the intended changes. I am fine with any of the first 4 options, but I must also say that this it outside of my comfort zone in terms of knowledge.

If there's no more feedback, what is it needed to unblock this contribution? If there's feedback, what is it needed to have a productive conversation that moves this forward instead of letting it stall?

isuruf:
@h-vetinari, my comment was not towards you, but for the comment highlighted.

I apologize if that seemed disrespectful. We've discussed this multiple times in Github and in core meetings and it seems like these discussions with you never seem to end. I stop responding for my own sanity.

As I said in the core meeting, I would prefer option 3, but okay with option 1.

beckermr:
OK thanks @isuruf! We have four core members who are happy with either 3 or 1 with one person having a preference for 3. @h-vetinari I think that is enough to go ahead with 3. However, you are doing the work so your judgement matters too.

h-vetinari:
> @isuruf: We've discussed this multiple times in Github and in core meetings

I've accommodated every feedback, the latest having just been "write up the plan we agreed to". Then crickets.

> @isuruf: it seems like these discussions with you never seem to end. I stop responding for my own sanity.

Now imagine having both a blocker and a pertinent question about solving it (e.g. something unclear), and it becomes an bottomless pit to throw energy into. Try to ask for clarification again a few weeks later - still silence. Try to move forward somehow - up pops the blocker again, mostly with no further elaboration.

> @beckermr: I think that is enough to go ahead with 3. However, you are doing the work so your judgement matters too.

Option 3. was proposed before three unanswered comments of mine, and until someone makes a convincing argument why it should be `-devel`, my point is that it increases balkanization of the `dev` / `-dev` / `-devel` space, rather than reduce it (e.g. all the recently added cuda libs use `-dev`; [ref](https://github.com/search?q=repo%3Aconda-forge%2Ffeedstock-outputs+%2F-dev%5Cb%2F&type=code); for `-devel` we have some archived CDT's, perl, blas & mkl).

I'm fine with the equivalent(-up-to-naming) option 4. though; with preference to name it `libboost-headers` until someone tells me why we should chop off the `lib` for the headers.

beckermr:
Sure! Option 1 is on the table too right?

h-vetinari:
> Sure! Option 1 is on the table too right?

Yes, that's the closest to what was discussed in the core call. I actually don't like keeping the old names (even though it's pragmatic vis-à-vis CFEP-20), so I have a slight preference for 4.

beckermr:
I think we shouldn't let the perfect be the enemy of the good. Whatever we do, the new system will be better than what we have now. I'm fine with 4 too.

h-vetinari:
So boost 1.83 was just released, and we should use this new version to make a clean break (so not merge any migrator PRs for 1.83). I also just found out about the [`libboost-mpi`](https://github.com/conda-forge/libboost-mpi-feedstock) feedstock (which no-one mentioned so far...), but at least that seems to fit into the naming scheme.

Concretely (following Option 4):
```
libboost{,-dev,-headers,-mpi,-python,-python-dev}  # {-static,-python-static} as possibile future extensions
```

I'll try to update the PR for that soon.

beckermr:
I didn't even know we had that feedstock. 🤦‍♂️

h-vetinari:
I looked at the content of `libboost-mpi`, and it inentionally does not contain the headers (they're deleted in the build script), only:
```
lib/cmake/boost_mpi-1.82.0/libboost_mpi-variant-shared.cmake
lib/libboost_mpi.so
lib/libboost_mpi.so.1.82.0
```

So that seems to fit well with the proposed setup here. @sdebionne please let us know if you think otherwise.

h-vetinari:
Also, I made a sweep for other boost flavours still lurking, and found [`boost_mp11`](https://github.com/conda-forge/boost_mp11-feedstock/blob/main/recipe/meta.yaml), which says:
```
    # Cannot be installed at the same time as boost-cpp as it clobbers the mp11 headers
    - boost-cpp <0a0
```
That would presumably become obsolete with the new `libboost-headers`, as `boost_mp11` contains no libs; but then I don't understand why it needed to be a separate package... it has no users in conda-forge AFAICT. CC @conda-forge/boost_mp11

jakirkham:
Would also be interested to hear from @jschueller and @xhochy who have been pretty active in both `boost-cpp` & `boost`

h-vetinari:
@xhochy is aware of this PR. We've had irregular offline conversations (on other things) and I've invited him several times to chime in here over the last months. He's a busy man though.

xhochy:
I've been on leave for four weeks and still catching up.I'll give this a re-read on Monday!

h-vetinari:
OK, the refactor into `libboost-dev` etc. is now finally getting somewhere. Unix should be good, probably still overlooked something on windows somewhere.

I rebased out the separate label again, as people seem to be in agreement that this should be the actual design now.

Beyond that, I see two questions:
* [ ] do we want to keep the compatibility outputs (`boost-cpp` & `boost`), or do we want a "clean cut"?
* [ ] do we want to migrate for 1.82 or 1.83?

I also see something weird on all the builds that finished so far.
```
SafetyError: The package for libboost-dev located at /home/conda/feedstock_root/build_artifacts/pkg_cache/libboost-dev-1.82.0-h00ab1b0_2
appears to be corrupted. The path 'lib/cmake/boost_headers-1.82.0/boost_headers-config.cmake'
has an incorrect size.
```

h-vetinari:
Aside from the points mentioned in the last comment, this should be good! The entire diff is probably a bit cumbersome to review, I think it might be easiest to go commit-by-commit after the merge of the different histories. Alternatively, [here's](https://github.com/h-vetinari/boost-feedstock/compare/merged_histories..unify) a collected diff from the merge until the head of the branch.

beckermr:
I think we should keep the compat outputs, but hint/lint and have a bot migrator.

beckermr:
I'd prefer 1.82 over 1.83 to stick with every other but I don't have strong feelings.

h-vetinari:
> I think we should keep the compat outputs, but hint/lint and have a bot migrator.

The renaming migrator is essentially ready (modulo renames). And we recently made it easier to add such rules to the linter too, so no problem there.

xhochy:
> I also see something weird on all the builds that finished so far

This sounds to me that a later build step is overwritting the contents of this file. As installed files are hardlinked globally, this has an effect on all subsequent installs.

xhochy (approved):
I'm happy with the changes.

Once the safetyerror is fixed, from my side, we can merge. 

I have a small preference for 1.82 because we have been using even releases all the time and it also leaves us with a backup release to use if there is a not-so-small cleanup to be done afterwards if we would have not taken everything into account correctly.

xhochy:
Slight preference for `-dev` from my side as this is what I'm used from Debian-based systems.

h-vetinari:
For me the point is about consistency. We have more (and more recent) packages with `-dev` than with `-devel`. I'd like us to become more consistent, not less, though which variant is less important to me.

The remaining slight preference for `-dev` is that the extra characters of `-devel` do not buy us any clarity IMO (I don't find parallels to other distros all that relevant, because our setup doesn't match in any case; e.g. we don't include static libs & don't have more granular boost libs).

> Before merging, let's discuss option 3 vs 4 in the next meeting.

Could you write down your argument here, so we don't take too much of the scarce core call time with a pretty bikesheddy topic?

isuruf:
1. All the packages in defaults are `-devel`. For eg: `tbb-devel, mkl-devel, openblas-devel`. I doubt defaults will go towards adding `-dev` simply because they only have `-devel`. If that's the case `-dev` will permanently diverge conda-forge from defaults. Correct me if I'm wrong @chenghlee.
2. CDTs all have `-devel`. This will be the case in the future as well if we stay with a centos based distro, but if we switch, this might change.
3. We already have `-devel` and `-dev`. It's better to pick one. (CUDA package naming have usually been done offline and it's hard to change them due to issues outside the control of c-f like nvidia legal. I would simply disregard the precedent set there as it's not done with community input)

h-vetinari:
Thanks. That reasoning is fine by me. If the others agree, I can change to `-devel` directly.

h-vetinari:
Given @xhochy's 👍, I've updated the PR.

recipe/build-py.sh
```diff
@@ -0,0 +1,97 @@
...
+# clean up directory from build.sh and reuse b2 built there
+rm -rf temp_prefix
+
+# $PREFIX/lib/cmake/boost_headers-$PKG_VERSION/boost_headers-config.cmake has
+# already been packaged (including the python headers!) into libboost-headers,
+# however the installation below will overwrite it with another (smaller) set
+# of metadata, causing a potentially corrupted package. Save the file in
+# question and restore it below.
+mkdir -p $SRC_DIR/cf_cmake
+cp $PREFIX/lib/cmake/boost_headers-$PKG_VERSION/boost_headers-config.cmake $SRC_DIR/cf_cmake/
```
xhochy, line 62: Can you try moving too here?

h-vetinari:
Thanks @xhochy, that fixed the `SafetyError`!

@isuruf, now that we've switched to your proposed `-devel`, do you still want to wait for the core call next week?

isuruf:
@conda-forge-admin, rerender

isuruf:
Works for me. If anyone else wants to discuss in the core meeting, that's fine, but not needed for me.

h-vetinari:
Do you want to keep the artefact persistence? (presumably that was just for double-checking the new builds...)

Otherwise I can revert (or rebase) this out. Are there any remaining points from your side before we merge this?

h-vetinari:
Thanks for the review!

> I'm a little surprised about the `lib`-prefix everywhere

I'm aware that this is not "standard" distro-style, but IMO we don't have to follow the unix naming conventions here. I'd rather have a consistent set of outputs (that appear next to each other in the alphabetically sorted `conda list`), not least because `libboost-headers` will be used like a library (for those packages that don't need more than the headers), and I think it'd be good to converge [towards](https://github.com/conda-forge/conda-forge.github.io/issues/1073) as few library patterns as possible.

We also didn't get to discuss `-dev` vs. `-devel` in the core call (I saw that @jaimergp had collected some extra numbers), I guess we can add it to the bikeshedding pile, if we want to wait for the next core call in two weeks.

recipe/build-py.sh
```diff
@@ -0,0 +1,97 @@
...
+elif [[ "${target_platform}" == linux* ]]; then
+    TOOLSET=gcc
+fi
+
+# http://www.boost.org/build/doc/html/bbv2/tasks/crosscompile.html
+cat <<EOF > ${SRC_DIR}/tools/build/src/site-config.jam
+using ${TOOLSET} : : ${CXX} ;
+EOF
+
+LINKFLAGS="${LINKFLAGS} -L${LIBRARY_PATH}"
```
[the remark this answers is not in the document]
h-vetinari: This is unchanged from the current recipe (I know that's hard to follow after the merge):
https://github.com/conda-forge/boost-feedstock/blob/e09a8e2f040b0223ee679421c70095698accd18a/recipe/build.sh#L33

The [commit](https://github.com/conda-forge/boost-feedstock/commit/a36f8a643fb3ae732060c358dcb528c5eb8ccbf0) that introduced that is pretty old, so I guess we can try to simplify this
```suggestion
LINKFLAGS="-L${LIBRARY_PATH}"
```

recipe/build-py.sh
```diff
@@ -0,0 +1,97 @@
...
+# clean up directory from build.sh and reuse b2 built there
+rm -rf temp_prefix
+
+# $PREFIX/lib/cmake/boost_headers-$PKG_VERSION/boost_headers-config.cmake has
+# already been packaged (including the python headers!) into libboost-headers,
+# however the installation below will overwrite it with another (smaller) set
+# of metadata, causing a potentially corrupted package. Save the file in
+# question and restore it below.
+mkdir -p $SRC_DIR/cf_cmake
+mv $PREFIX/lib/cmake/boost_headers-$PKG_VERSION/boost_headers-config.cmake $SRC_DIR/cf_cmake/
```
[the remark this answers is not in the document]
h-vetinari: I haven't diffed the files. Though the headers should be independent of the python version, and the global `build.sh` that creates the original `boost_headers-config.cmake` does have the python bits enabled and should be a proper superset of what the python-only build populates in this file.

recipe/build.sh
```diff
@@ -32,17 +28,20 @@ EOF
...
+CFLAGS="$(echo ${CFLAGS} | sed 's/ -march=[^ ]*//g' | sed 's/ -mcpu=[^ ]*//g' |sed 's/ -mtune=[^ ]*//g')" \
+    CXX=${CXX_FOR_BUILD:-${CXX}} CC=${CC_FOR_BUILD:-${CC}} ./bootstrap.sh \
     --prefix="${PREFIX}" \
     --with-toolset=${TOOLSET} \
     --with-icu="${PREFIX}" \
     --with-python="${PYTHON}" \
-    --with-python-root="${PREFIX} : ${PREFIX}/include/python${PY_VER}m : ${PREFIX}/include/python${PY_VER}" \
-    2>&1
+    --with-python-root="${PREFIX} : ${PREFIX}/include/python${PY_VER}" \
+    || (cat bootstrap.log; exit 1)
```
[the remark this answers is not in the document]
h-vetinari: This comes from [boost-cpp](https://github.com/conda-forge/boost-cpp-feedstock/blob/f6b34b392599fe6aa4ef48a679a90d289e5f646e/recipe/build.sh#L38).

Clarity is clearly pretty subjective (I'm a noob in bash, but your suggestions aren't particularly clearer to me[^1]), but I think the intent is pretty clearly: "in case of failure, print log, _then_ fail". Would the following be a less ambiguous expression of that?
```suggestion
    || (cat bootstrap.log && exit 1)
```

[^1]: I wouldn't have guessed that `false` fails the build, and wouldn't know how `{ }` is different from `( )` here.

recipe/install-lib.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -ex
+
+if [[ "$PKG_NAME" == "libboost-headers" ]]; then
+    # for libboost-headers, only the headers;
+    # since we have no host dependencies, the folder $PREFIX/include does not exist yet,
+    # so the syntax for `copy` works differently than below (where $PREFIX/lib exists).
+    cp -R temp_prefix/include $PREFIX/include
```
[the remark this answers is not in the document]
h-vetinari: I did actually check the man-pages when I was trying to figure out why the `/*` didn't work as expected; thanks for the tip & explanation!

h-vetinari:
@sdebionne brought up the mpi headers again in https://github.com/conda-forge/boost-cpp-feedstock/issues/131. I wanted to bring this up for a wider audience here to ensure we're not overlooking something.

> [OP of that issue] @jakirkham: Currently the MPI headers are getting packaged in `boost-cpp`. However these should be in `libboost-mpi` ( https://github.com/conda-forge/libboost-mpi-feedstock/issues/9 ). These should be dropped from `boost-cpp` and `libboost-mpi` rebuilt.

> [snip]

> @sdebionne: With the boost unification effort, I guess this can be safely closed?

This matches my understanding that the intention now is for `libboost-headers` to contain all headers; in this case, we're not even changing the status quo, as the most recent builds of `libboost-mpi` only contain
```
lib/cmake/boost_mpi-1.82.0/libboost_mpi-variant-shared.cmake
lib/libboost_mpi.so
lib/libboost_mpi.so.1.82.0
```
and need `boost-cpp` for the headers. In the context of https://github.com/conda-forge/libboost-mpi-feedstock/issues/9, it should then be possible for that feedstock to just rely on `libboost-headers`.

jaimergp:
> We also didn't get to discuss `-dev` vs. `-devel` in the core call (I saw that @jaimergp had collected some extra numbers), I guess we can add it to the bikeshedding pile, if we want to wait for the next core call in two weeks.

Yes, I searched `feedstock-outputs` for `*-dev.json` and `*-devel.json`

- [-dev suffix](https://github.com/search?q=repo%3Aconda-forge%2Ffeedstock-outputs+path%3A%22-dev.json%22&type=code): 33 packages
- [-devel suffix](https://github.com/search?q=repo%3Aconda-forge%2Ffeedstock-outputs+path%3A-devel.json&type=code): 10 packages

Maybe there's a better regex there (e.g. what about devel infixes), but you get the idea: dev is 3x more popular than devel, but we are just talking about 33 vs 10 packages, so no too many anyway.

h-vetinari:
> but you get the idea: dev is 3x more popular than devel

I just did a quick count and 17 out of the 33 are for the recent nvidia packages that Isuru addressed in his comment, so once you subtract those, the balance is _relatively_ even. I still have a slight preference for `-dev` but the consistency with Anaconda is the strongest argument against that for me.

In any case, I wanted to bring this up in the core call less for the particular case of boost here, but more about taking a conda-forge wide decision (or at least direction) which one we prefer going forward.

jaimergp:
I added data about infixes too. Most of the devel stuff is either a CDT or a Perl package, but I guess that counts too. That said, you are right, and this is conda-forge-wide issue, not just Boost. Maybe worth discussing naming patterns separately: what our conventions are and perhaps even more importantly how to enforce them.

h-vetinari:
We discussed the naming in the core call today, seems that everyone's fine to move forward with `-devel`. This PR is IMO ready to merge, and I'd like to ask for any final thoughts/concerns @conda-forge/core @conda-forge/boost @conda-forge/boost-cpp

h-vetinari:
> This PR is IMO ready to merge, and I'd like to ask for any final thoughts/concerns

Alright, it's been almost a week, let's get this in. Thanks everyone!

PR #172, opened 2023-09-27
Build windows against icu

### LINKED ISSUE #168: Why no ICU on windows?
How are you able to write portable code using boost if you don't build it using the same dependencies?

h-vetinari:
Trying to make the build more uniform, see #168, following instructions from [upstream](https://www.boost.org/doc/libs/1_83_0/libs/locale/doc/html/building_boost_locale.html#bb_build_opts) (doesn't appear to have the same b2-interface like the other libs).

Plus some minor clean-ups (which I avoided for unification to keep the diff minimal).

Closes #168


h-vetinari:
This seems to be running into something vaguely similar to https://github.com/bfgroup/b2/issues/189, as we're getting:
```

    - has_icu builds           : no [2]
    - iconv (libc)             : no [2]
    - iconv (separate)         : yes [2]
    - icu                      : no [2]
```

h-vetinari:
Still nada:
```
    - icu                      : no [2]
    - iconv (libc)             : no [2]
    - iconv (separate)         : yes [2]
```

sdebionne:
Maybe it is worth to try to set `ICU_PATH` to %LIBRARY_PREFIX% ?

recipe/bld.bat
```diff
@@ -24,6 +24,10 @@ mkdir temp_prefix
     threading=multi ^
     link=shared ^
     cxxstd=20 ^
+    boost.locale.icu=on ^
+    boost.locale.std=on ^
+    boost.locale.winapi=on ^
+    -s ICU_PATH=%LIBRARY_PREFIX% ^
```
h-vetinari, line 30: > Maybe it is worth to try to set `ICU_PATH` to %LIBRARY_PREFIX% ?

already done

h-vetinari:
No change.

Issue #176, opened 2023-10-01
Missing components showing up in dependent packages


h-vetinari:
In the boost unification (#164), we tried as much as possible - aside from the output reshuffling - to keep the build exactly as before. In particular, the build script is taken essentially verbatim from the previous [boost-cpp](https://github.com/conda-forge/boost-cpp-feedstock/tree/main/recipe) (with the exception of adding the python build to have all the headers at once).

However, during the migration I'm now seeing cases of missing components, e.g. [here](https://github.com/conda-forge/rdkit-feedstock/pull/136)
```
-- Could NOT find Boost (missing: system serialization iostreams) (found suitable version "1.82.0", minimum required is "1.58.0")
```
or [here](https://github.com/conda-forge/vigra-feedstock/pull/123)
```
Could NOT find Boost (missing: thread system date_time chrono) (found suitable version "1.82.0", minimum required is "1.55.0")
```

These all sound very stdlib-related, so perhaps those feedstocks are defaulting to a too-old C++ version?

isuruf:
`libboost-python-devel` probably needs `libboost-devel` as a run time requirement. (And also a run_exports)

h-vetinari:
Ah yeah, that makes sense. PR coming up.

Next issue: [folly](https://github.com/conda-forge/folly-feedstock/pull/143) cannot find a header anymore,
```
In file included from /tmp/tmpp7qryo05/info/recipe/arena_test.c:1:
/home/conda/feedstock_root/build_artifacts/folly_1696034969559/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/include/folly/memory/Arena.h:25:10: fatal error: boost/intrusive/slist.hpp: No such file or directory
   25 | #include <boost/intrusive/slist.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

even though it's [clearly](https://github.com/boostorg/boost/blob/boost-1.82.0/.gitmodules#L391-L395) still part of boost.

h-vetinari:
> (And also a run_exports)

I'm not sure that's the best pattern here? What can happen is that a library really only depends on `libboost-python` at runtime, but expressing that accurately now requires:
```
build:
  ignore_run_exports:
    - libboost
```

[...]

PR #177, opened 2023-10-01
Ensure python devel package depends on `libboost-devel`

h-vetinari:
Addresses part of #176, plus some clean-ups I left out of the unification PR to not blow up the diff any further.


recipe/meta.yaml
```diff
@@ -285,7 +286,7 @@ outputs:
         - python
         - {{ pin_compatible('numpy') }}
         - {{ pin_subpackage("libboost-python", exact=True) }}
-        - {{ pin_subpackage("libboost-headers", exact=True) }}
+        - {{ pin_subpackage("libboost-devel", exact=True) }}
```
h-vetinari, line 289: @isuruf, I'm almost certain that run-depending in `libboost-devel` ensures that anyone host-depending on `libboost-python-devel` will in fact pick up the `libboost` run-export. If you want me to make that more explicit, I'm happy to do that.

isuruf: Not true.

h-vetinari: Huh, TIL - I though it picks all the run-exports from what's present in host (after resolution). In any case, I added the explicit run-export on `libboost` now.

Issue #178, opened 2023-10-01
Corrupted libs with osx-arm64 build?


h-vetinari:
Just noticed this in CI:
```
SafetyError: The package for libboost located at /Users/runner/miniforge3/pkgs/libboost-1.82.0-h72cdd8a_4
appears to be corrupted. The path 'lib/libboost_atomic.dylib'
has an incorrect size.
  reported size: 87312 bytes
  actual size: 86178 bytes

SafetyError: The package for libboost located at /Users/runner/miniforge3/pkgs/libboost-1.82.0-h72cdd8a_4
appears to be corrupted. The path 'lib/libboost_chrono.dylib'
has an incorrect size.
  reported size: 61328 bytes
  actual size: 60194 bytes

SafetyError: The package for libboost located at /Users/runner/miniforge3/pkgs/libboost-1.82.0-h72cdd8a_4
appears to be corrupted. The path 'lib/libboost_container.dylib'
has an incorrect size.
  reported size: 123552 bytes
  actual size: 122421 bytes
```

and so on...

This seems to have been overlooked in #164, unfortunately. 😑 

We did have some other [SafetyErrors](https://github.com/conda-forge/boost-feedstock/pull/164#issuecomment-1677358711) for the unification related to CMake metadata, which got fixed in 43373e65b5721f029fd59db95cf9ab7cd74dc9b8 [and](https://github.com/conda-forge/boost-feedstock/pull/164#discussion_r1297118599) 3c87dcf6d04df7b904c46d5267459d5a78a4624d; this could be similar in principle.

What I could imagine happening: we're generating signature with cctools upon first installation (for `libboost`), which changes the artefacts. Once we copy them again for `libboost-devel`, the signature doesn't get generated, and so conda gets confused with the sizes?

PR #179, opened 2023-10-01
fix SafetyErrors on osx-arm

h-vetinari:
Tentative fix for #178 🤞


h-vetinari:
Alrighty, that worked first try 🥳

jakirkham:
Is this due to `@rpath` fixing? Wonder if we should do that ourselves in the build so that conda-build's work would be a no-op

h-vetinari:
It's my understanding (as I documented in the script) that this is due to the signatures we add with cctools, which only seems to run for new files, and not for reinstalled ones.

PR #180, opened 2023-10-03
backport patch for removed-in-C++17 std::{binary,unary}_function (yet again)

h-vetinari:
Seems we get to relive https://github.com/conda-forge/boost-cpp-feedstock/issues/134 & https://github.com/conda-forge/boost-cpp-feedstock/pull/135, only slightly differently.

Still getting failures in https://github.com/conda-forge/graph-tool-feedstock/pull/114 of the kind:
```
In file included from ./src/graph/histogram.hh:26:
In file included from $PREFIX/include/boost/multi_array.hpp:34:
In file included from $PREFIX/include/boost/multi_array/multi_array_ref.hpp:32:
$PREFIX/include/boost/functional.hpp:45:24: error: no member named 'unary_function' in namespace 'std'; did you mean '__unary_function'?
            using std::unary_function;
                  ~~~~~^
```

That's because apparently, we're not automatically picking up `_HAS_AUTO_PTR_ETC`, which is the only [guard](https://github.com/boostorg/functional/blob/boost-1.82.0/include/boost/functional.hpp#L24) from using the deprecated functions as of boost 1.82.

This got fixed (presumably the last time) in https://github.com/boostorg/functional/commit/6a573e4b8333ee63ee62ce95558c3667348db233, so backport that patch.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -21,9 +21,11 @@ source:
   patches:
     # ensure our compiler flags get used during bootstrapping 
     - patches/0001-Add-default-value-for-cxx-and-cxxflags-options-for-t.patch
+    # backport https://github.com/boostorg/functional/commit/6a573e4b8333ee63ee62ce95558c3667348db233
+    - patches/0001-Define-unary_function-and-binary_function-unconditio.patch
 
 build:
-  number: 5
+  number: 6
 
 requirements:
   build:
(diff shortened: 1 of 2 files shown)


h-vetinari:
@conda-forge/boost could someone please review? Given that this commit landed in boost 1.83, I think we should be safe to backport it, but would appreciate another set of eyes.

CC @conda-forge/graph-tool @traversaro

PR #206, opened 2024-05-14
also test for charconv library

sdebionne:
Adding `charconv` to the list of build libraries.

Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -30,7 +30,7 @@ source:
     - patches/0001-Merge-pull-request-183-from-boostorg-ppc64le.patch
 
 build:
-  number: 1
+  number: 0
   script_env:
     - PY_DUMMY_VER={{ PY_DUMMY_VER }}
     - NP_DUMMY_VER={{ NP_DUMMY_VER }}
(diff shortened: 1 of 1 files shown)


recipe/meta.yaml
```diff
@@ -30,7 +30,7 @@ source:
     - patches/0001-Merge-pull-request-183-from-boostorg-ppc64le.patch
 
 build:
-  number: 0
+  number: 1
```
h-vetinari, line 33:
```suggestion
  number: 0
```

h-vetinari (commented):
The title is a bit misleading. The only place where we're "adding" charconv is in testing that the required (shared) libraries are there at the end of the build.

This can be seen also since the 1.85 from #196 were failing specifically in charconv already before this PR (and the patch to fix it). Long story short, this is a reasonable test improvement, but not worth a separate build (so we can just leave the build number as-is).

recipe/meta.yaml
sdebionne: Fine with me, I was just blindly following the checklist.

h-vetinari (approved):

PR #223, opened 2024-12-17
Add libbacktrace and enable boost-stacktrace.

oursland:
Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)


recipe/meta.yaml
```diff
@@ -4,7 +4,7 @@
     "atomic", "charconv", "chrono", "cobalt", "container", "context", "contract", "coroutine",
     "date_time", "filesystem", "graph", "iostreams", "locale", "log", "log_setup",
     "math_c99", "math_c99f", "math_tr1", "math_tr1f", "prg_exec_monitor",
-    "program_options", "random", "regex", "serialization", "system", "thread",
+    "program_options", "random", "regex", "serialization", "stacktrace", "system", "thread",
     "timer", "type_erasure", "unit_test_framework", "wave", "wserialization"
 ] %}
 {% set boost_libs = boost_libs + ["math_c99l", "math_tr1l"] %}  # [not ppc64le]
```
h-vetinari, line 10: You cannot add this to the overall `boost_libs` because these are expected to exist on all platforms (and you're not enabling it on windows). Could be something like

```suggestion
    "program_options", "random", "regex", "serialization", "system", "thread",
    "timer", "type_erasure", "unit_test_framework", "wave", "wserialization"
] %}
# there's no backtrace built on windows in conda-forge yet
{% set boost_libs = boost_libs + ["stacktrace"] %}              # [not win]
{% set boost_libs = boost_libs + ["math_c99l", "math_tr1l"] %}  # [not ppc64le]
```

oursland:
Dang.  I put this work in hoping to get Boost.Stacktrace with support of `stacktrace_from_exceptions` on macOS, but apparently it only builds on Linux and Windows as `libc++`-based systems result in a memory leak.

I am uncertain if this PR is worth pursuing.  It brings in `libbacktrace` by hacking in static linking in the `Jamfile.v2` on the `#[unix]` platforms.

h-vetinari:
> but apparently it only builds on Linux and Windows as `libc++`-based systems result in a memory leak.

Where does this conclusion/information come from? 

> I am uncertain if this PR is worth pursuing. It brings in `libbacktrace` by hacking in static linking in the `Jamfile.v2` on the `#[unix]` platforms.

Yeah, we don't want to do static linkage. Why couldn't it work with the regular shared approach?

oursland:
> Where does this conclusion/information come from?

https://github.com/boostorg/stacktrace/blob/develop/src/from_exception.cpp#L167

> Yeah, we don't want to do static linkage. Why couldn't it work with the regular shared approach?

The executables do not have a `LC_RPATH` set and cannot resolve the location of `libbacktrace.dylib`.  I'm certain enough hacking of the `Jamfile.v2` would resolve the issue, but unfortunately the aforementioned limitation makes this less useful than desired.

I'm looking at making a `conda-forge` recipe for [`cpptrace`](https://github.com/jeremy-rifkin/cpptrace), which does not appear to suffer the same limitations.

mike-jn:
Hi,

I wasn't able to get boost-stacktrace to build in my cmake project -- does this PR mean that the conda-forge build of boost is simply missing it?

h-vetinari:
Hi @mike-jn; indeed we don't build with backtrace support at the moment.

This is from the logs of the most recent linux build
```
    - boost.stacktrace.addr2line : yes [2]
    - boost.stacktrace.backtrace : no [2]  # <--- !!!
    - boost.stacktrace.backtrace : no [4]  # <--- !!!
    - boost.stacktrace.basic   : no [2]
    - boost.stacktrace.from_exception : yes
    - boost.stacktrace.from_exception : yes
    - boost.stacktrace.windbg  : no [2]
    - boost.stacktrace.windbg  : no [4]
    - boost.stacktrace.windbg_cached : no [2]
    - boost.stacktrace.windbg_cached : no [4]
[...]
[2] gcc-13/release/x86_64/cxxstd-20-iso/python-3.10/threading-multi/visibility-hidden
[3] gcc-13/release/x86_64/cxxstd-20-iso/link-static/python-3.10/threading-multi/visibility-hidden
[4] gcc-13/release/x86_64/build-no/cxxstd-20-iso/python-3.10/threading-multi/visibility-hidden
```

This is not as much a conscious choice as a reflection of the fact that the build system checks for the presence of `backtrace` (which we simply hadn't added to the dependencies here because it's not necessary otherwise), and no-one (before @oursland) ran into the issue of needing the backtrace portion from boost.

PRs as welcome!

PR #229, opened 2025-03-16
boost v1.88.0

h-vetinari:
~On top of #228~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,8 @@
 /build_artifacts
 
 *.pyc
+
+# Rattler-build's artifacts are in `output` when not specifying anything.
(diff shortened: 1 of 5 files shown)


hmaarrfk:
we should start a migrator right?

h-vetinari:
> we should start a migrator right?

I wasn't in a hurry on this one, but I'm also not against it. 1.86 was more breaking than usual, so I was thinking of doing this as soon as a reason arises (e.g. some project needs it), and leaving a bit more time for the ecosystem to adapt. Feel free to propose a migrator though!

hmaarrfk:
Ok understood.

PR #269, opened 2026-08-29
Make libboost-headers noarch

ClayWarren:
Publishes separate noarch generic header variants for Unix and Windows, constrained by __unix and __win. This makes the Windows headers available on win-arm64 without building the compiled Boost outputs there.


isuruf (changes requested):
Just make libboost-headers package noarch: generic with __unix and __win variants.

ClayWarren:
Updated as suggested: libboost-headers is now noarch generic with __unix and __win variants built on linux_64 and win_64. The native win-arm64 provider and architecture-specific changes are removed.

h-vetinari:
Please send a matching PR to `main` as well.

PR #270, opened 2026-08-30
Make libboost-headers noarch on main

ClayWarren:
Matches #269 on `main` for Boost 1.92.x.

- makes `libboost-headers` noarch
- preserves Unix/Windows constraints with `__unix` and `__win`


h-vetinari:
Thanks!

